PR #6, opened 2019-06-07
Detox tests added

DanielSanudo:
# Summary

- Detox tests added for iOS and Android
- Example app doesn't require anymore its own package.json
- Example app dependencies now reference to the root node_modules

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

- [X] I mentioned this change in `CHANGELOG.md`


example/android/build.gradle
```diff
@@ -2,10 +2,11 @@
 
 buildscript {
     ext {
+        kotlinVersion = "1.3.0"
         buildToolsVersion = "28.0.3"
-        minSdkVersion = 16
+        minSdkVersion = 18
         compileSdkVersion = 28
-        targetSdkVersion = 28
```
Swaagie: why did this need to be downgraded?

example/e2e/init.js
```diff
@@ -0,0 +1,25 @@
+const detox = require('detox');
+const config = require('../../package.json').detox;
+const adapter = require('detox/runners/jest/adapter');
+const specReporter = require('detox/runners/jest/specReporter');
+
+// Set the default timeout
+jest.setTimeout(120000);
+jasmine.getEnv().addReporter(adapter);
```
Swaagie: where is `jasmine` coming from here? Does detox provide that on `global` by default?

example/App.js
```diff
@@ -34,21 +34,40 @@ export default class App extends Component<Props> {
     this.show('time');
   }
 
+  mmddyyyy = (date) => {
+    var yyyy = date.getUTCFullYear();
+    var mm = date.getUTCMonth() < 9 ? '0' + (date.getUTCMonth() + 1) : (date.getUTCMonth() + 1); // getUTCMonth() is zero-based
```
Swaagie: This can probably be cleaned up a bit if the optional `0` is prepended after the actual, also would this display `010` if the returned month is `9`? Also would prefer regular functions here as part of the `class`

example/android/build.gradle
DanielSanudo: This is what fixed the issue for running the detox tests in android-release. When using `targetSdkVersion = 28` the app was getting installed but the tests weren't running.

example/e2e/init.js
3rd-Eden: `jest` is actually using `jasmine` so it's coming from `jest/adapter`

package-lock.json
```diff
@@ -766,9 +766,9 @@
       },
       "dependencies": {
         "core-js": {
-          "version": "3.1.2",
-          "resolved": "https://artifactory.secureserver.net/artifactory/api/npm/node-virt/core-js/-/core-js-3.1.2.tgz",
-          "integrity": "sha1-JUmiz7PKGl2FHJ94OOiygs7y87o=",
+          "version": "3.1.3",
+          "resolved": "https://artifactory.secureserver.net/artifactory/api/npm/node-virt/core-js/-/core-js-3.1.3.tgz",
```
3rd-Eden: please use the `npm.registry`

example/App.js
3rd-Eden: As this is an exampe app, can't we just use `data-fn`'s to clean this up a little bit?

example/babel.config.js
```diff
@@ -1,13 +0,0 @@
-module.exports = function (api) {
```
3rd-Eden: Does the normal example still work after you deleted all of this?

example/App.js
```diff
@@ -72,4 +91,8 @@ const styles = StyleSheet.create({
     fontSize: 20,
     fontWeight: 'bold',
   },
+  dateTimeText: {
+    fontSize: 16,
+    fontWeight: 'normal',
+  },
```
3rd-Eden: Remove extra `,` -- but that also means that a linter didn't run, so can you run the linter over these files?

example/babel.config.js
DanielSanudo: Yep, you just need to do `npm run start` and then `npm run start:ios` or `npm run start:android`. All the dependencies are now pointing to the root `node_modules`.

example/App.js
DanielSanudo: I have included the `example` folder for linting and it is requiring that `,`. I guess that this is how it is defined in `@react-native-community`.

example/e2e/init.js
Swaagie: gotta luv test globals 😁

example/App.js
Swaagie: This is part of the react-native-community eslint config, they like their comma's :)

example/App.js
DanielSanudo: I have ended up using `moment`. `date-fns` was giving me problems for timezones and it seems that there is no way of defining a timezone for the format function.

PR #80, opened 2020-01-03
Support compiling with other platforms

aikewoody:
# Summary
When using this plugin in a project which targets an environment other than Android or iOS it can't be compiled because the file `datetimepicker` has no default entry.

## Test Plan
```
npm install --save git+https://github.com/aikewoody/react-native-datetimepicker.git#feature-support-all-platforms
# Install react-native-web
# If you're inside an Expo project: https://docs.expo.io/versions/v33.0.0/introduction/running-in-the-browser/
expo install react-native-web react-dom
expo start --web
```

An Expo project with the `react-native-community/react-native-datetimepicker` plugin installed.

```
npm install --save @react-native-community/datetimepicker@latest
# Install react-native-web
# If you're inside an Expo project: https://docs.expo.io/versions/v33.0.0/introduction/running-in-the-browser/
expo install react-native-web react-dom
expo start --web
```

The following error appears:
```
web  Failed to compile.
/Users/.../node_modules/@react-native-community/datetimepicker/src/index.js
Module not found: Can't resolve './datetimepicker' in '/Users/.../node_modules/@react-native-community/datetimepicker/src'
```

## Compatibility
| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Web |    ✅     |
| macOS |    ✅     |
| Windows |    ✅     |

## Checklist

### DIFF
--- a/src/datetimepicker.js
+++ b/src/datetimepicker.js
@@ -1,0 +1,1 @@
+export default undefined;
--- a/src/datetimepicker.js
+++ b/src/datetimepicker.js
@@ -1,3 +1,6 @@
+import React from 'react';
+import { Platform } from 'react-native';
+
 export default function DateTimePicker() {
   React.useEffect(() => {
     console.warn(`DateTimePicker is not supported on: ${Platform.OS}`);
--- a/src/datetimepicker.js
+++ b/src/datetimepicker.js
@@ -1,1 +1,6 @@
-export default undefined;
+export default function DateTimePicker() {
+  React.useEffect(() => {
+    console.warn(`DateTimePicker is not supported on: ${Platform.OS}`);
+  }, [])
+  return null;
+}
--- a/src/index.js
+++ b/src/index.js
@@ -1,4 +1,3 @@
 import RNDateTimePicker from './datetimepicker';
 
 export default RNDateTimePicker;
-


chenguzhen87:
expo start -- web, I also encounter this problem, how to solve it
image

aikewoody:
> expo start -- web, I also encounter this problem, how to solve it
> image

Your problem can be fixed by using the repo and branch from this pull request (https://github.com/aikewoody/react-native-datetimepicker/tree/feature-support-all-platforms):
`npm install --save git+https://github.com/aikewoody/react-native-datetimepicker.git#feature-support-all-platforms`

Or by locally fixing `react-native-datetimepicker` in your `node_modules`:
echo 'export default undefined;' >> 'node_modules/@react-native-community/datetimepicker/src/datetimepicker.js'

aikewoody:
> Hi, thanks for the PR! The native files export a function (react component), wouldn't it make sense to export a function (a function that just renders nothing) instead of undefined?

@vonovak Good point using an empty function would prevent an error because of rendering an undefined value. The drawback is that checking for a truthy value in this way is not longer possible:
`{DateTimePicker && <DateTimePicker onChange={this.onChange} value={date} />}`

Instead users on web (or any other platform besides Android and iOS) need to use this check:
`{Platform.OS === 'android' || Platform.OS === 'ios' ? <DateTimePicker onChange={this.onChange} value={date} /> : null}`

Given this new insight what do you prefer?

aikewoody:
@EvanBacon Thanks for your insight, I really like your solution by warning the user about it.

vonovak:
Thank you both!

PR #106, opened 2020-02-03
added a line to the RNDatePickerDialogFragment.java file to stop the library from breaking on android

kamsy:
# Summary
I installed the library recently for an app I was building and it crashed on Android, read the loads and fixed the bug, so I'm basically creating this PR to help others and myself for a later date when I'd reach for this library

## Test Plan

Error:
image

npx react-native run-android

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist


vonovak:
hi, thanks for the PR. The method should never return null, so used https://github.com/react-native-community/react-native-datetimepicker/pull/107 instead

kamsy:
That’s odd. I just tried out the fix of #107 and my app crashed again.

vonovak:
@kamsy that's odd indeed. Can you share the stacktrace?

PR #115, opened 2020-02-11
Fix android time picker returning today's date instead of the given date

### LINKED ISSUE #94: Wrong value returned from picker
# Bug

Time picker returns today's date instead of given date.

In `datetimepicker.android.js` 53rd line should be `const date = new Date(value);` instead of `const date = new Date();`

roconnorr:
# Summary

After a time value is chosen in the android picker, use the value prop to create the date object so it doesn't default to returning the current date when using the time picker. Closes #94

A test project for android with a time picker set to ```mode="time"```

Pass the time picker a date that is not today, log the value returned from the picker after selecting a time, and see that it returns today's date.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

- [ ] I added the documentation in `README.md` - not applicable for bug fix
- [ ] I updated the typed files (TS and Flow) - not applicable for bug fix
- [ ] I added a sample use of the API in the example project (`example/App.js`) - not applicable for bug fix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+### 2.2.2
+- Fix android time picker returning today's date instead of the given date
+
 ### 2.2.0
 
 - Fix podspec to get source from tag [#103](https://github.com/react-native-community/react-native-datetimepicker/pull/103)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # CHANGELOG
 
-### 2.2.2
+### 2.2.1
 - Fix android time picker returning today's date instead of the given date
 
 ### 2.2.0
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+### 2.2.2
```
luancurti: Can you change to `2.2.1`?

roconnorr: Yup I have changed it, looks like there is already a tag in the repo for 2.2.1 though?

luancurti: You are right, I'll fix this in another PR thanks

kptac:
This issue has not been fixed still. Using the latest v 2.3.0, On setting time, the date still changes to the current date.

lander854:
same here

kptac:
@roconnorr - kindly look into this issue again. The issue is still not resolved. Thanks

PR #120, opened 2020-02-28
Add styling prop for iOS picker

lochness42:
# Summary

At the moment on iOS picker has fixed height of 216px, this PR exposes a new style property which is passed down to picker itself (style property is only propagated to internal View container and picker can overflow it)
If this style is not defined, we will fall back to height of 216px so it won't break existing apps.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     	|
| Android |    ❌     	|

## Checklist

### DIFF
--- a/src/datetimepicker.ios.js
+++ b/src/datetimepicker.ios.js
@@ -69,7 +69,7 @@ export default class Picker extends React.Component<IOSNativeProps> {
       mode,
       minuteInterval,
       timeZoneOffsetInMinutes,
-      iOsPickerStyle,
+      iOsPickerStyle = {},
     } = this.props;
 
     invariant(value, 'A date or time should be specified as `value`.');
@@ -82,11 +82,7 @@ export default class Picker extends React.Component<IOSNativeProps> {
         <RNDateTimePicker
           testID={testID}
           ref={this._picker}
-          style={
-            iOsPickerStyle !== undefined && iOsPickerStyle !== null
-              ? iOsPickerStyle
-              : styles.picker
-          }
+          style={[styles.picker, iOsPickerStyle]}
           date={dates.value}
           locale={locale !== null && locale !== '' ? locale : undefined}
           maximumDate={dates.maximumDate}
--- a/src/datetimepicker.ios.js
+++ b/src/datetimepicker.ios.js
@@ -82,7 +82,11 @@ export default class Picker extends React.Component<IOSNativeProps> {
         <RNDateTimePicker
           testID={testID}
           ref={this._picker}
-          style={iOsPickerStyle || styles.picker}
+          style={
+            iOsPickerStyle !== undefined && iOsPickerStyle !== null
+              ? iOsPickerStyle
+              : styles.picker
+          }
           date={dates.value}
           locale={locale !== null && locale !== '' ? locale : undefined}
           maximumDate={dates.maximumDate}
--- a/src/datetimepicker.ios.js
+++ b/src/datetimepicker.ios.js
@@ -9,7 +9,7 @@
  * @format
  * @flow strict-local
  */
-import {View, StyleSheet} from 'react-native';
+import {StyleSheet} from 'react-native';
 import RNDateTimePicker from './picker';
 import {toMilliseconds} from './utils';
 import {MODE_DATE} from './constants';
--- a/src/datetimepicker.ios.js
+++ b/src/datetimepicker.ios.js
@@ -69,6 +69,7 @@ export default class Picker extends React.Component<IOSNativeProps> {
       mode,
       minuteInterval,
       timeZoneOffsetInMinutes,
+      iOsPickerStyle,
     } = this.props;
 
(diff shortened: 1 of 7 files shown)


src/datetimepicker.ios.js
```diff
@@ -81,7 +82,11 @@ export default class Picker extends React.Component<IOSNativeProps> {
         <RNDateTimePicker
           testID={testID}
           ref={this._picker}
-          style={styles.picker}
+          style={
```
luancurti: I think you should apply both styles if you can override the `style.picker` you can set the same `property` on `iOsPickerStyle`. With this implementation this comparison isn't necessary you can pass an array to style `style={[styles.picker, iOsPickerStyle]}` and on the line `72` you can set a default style with an empty object `iOsPickerStyle  = {}`

luancurti: What do you think? @lochness42

lochness42: My original thought was not to enforce overriding height when user just wants to use flex value. But using `height: 100%` isn't too bad so will use style array.

src/datetimepicker.ios.js
```diff
@@ -69,6 +69,7 @@ export default class Picker extends React.Component<IOSNativeProps> {
       mode,
       minuteInterval,
       timeZoneOffsetInMinutes,
+      iOsPickerStyle = {},
```
vonovak: I'd vote for this prop to be named just `style`. There are several other props such as `timeZoneOffsetInMinutes` that only work on one platform, and we never use any ios/android suffix/prefix to distinguish those.

lochness42: Do you suggest to use same styles on outer container and also DatePicker?

lochness42: Style is already used here - https://github.com/react-native-community/react-native-datetimepicker/blob/e6d2980230b1e27fa8aed989ccbb79897e6ef5eb/src/datetimepicker.ios.js#L80

vonovak: huh I didn't know that! (it's not documented). That brings me to a question - why is `RNDateTimePicker` wrapped in a View? Would it make sense to remove that wrapping view?

src/types.js
```diff
@@ -99,6 +100,15 @@ export type IOSNativeProps = $ReadOnly<{|
    * instance, to show times in Pacific Standard Time, pass -7 * 60.
    */
   timeZoneOffsetInMinutes?: ?number,
+
+  /**
+   * Styles for internal picker component
+   *
+   * By default picker has fixed height of 216 on iOS ignoring it's parent's container constraints
```
vonovak:
```suggestion
   * By default, picker has fixed height of 216 on iOS, ignoring its parent's container constraints
```

src/datetimepicker.ios.js
lochness42: I can remove it, I've tested it and it works without the external View.
I just wonder whether there will be breaking change in case someone had a hack to get around inability to change height of the datepicker.

lochness42: Alright added it to example to `App.js` and has documentation in `Readme.md`. I've kept it as **iOS only** due to the fact that you can't really style Android pickers this way.

vonovak: > I can remove it, I've tested it and it works without the external View.

in that case I'd remove it, and if folks need to wrap it, they can do it themselves

I checked https://github.com/facebook/react-native/blame/master/Libraries/Components/DatePicker/DatePickerIOS.ios.js and from the blame, it doesn't seem like there was specific reason for adding the wrapping view

lochness42: Ok done

test/index.test.js
```diff
@@ -74,10 +71,10 @@ describe('DatePicker', () => {
       .create(<DatePicker style={style} value={new Date(DATE)} />)
       .toJSON();
 
-    expect(tree).toHaveProperty(
-      'props.style.backgroundColor',
-      style.backgroundColor,
-    );
+    expect(tree).toHaveProperty('props.style', [
```
vonovak: the `it` label of this test no longer makes sense because there is no `View` wrapper, could you fix that? Thanks!

vonovak:
@lochness42 I'm also wondering - just because people may try to use various styling options - what are the limitations as to what styling props people can use here? I see the motivation was to change picker size. Is it possible to change background color or border for example? Thanks!

lochness42:
@vonovak background color and border does work
Simulator Screen Shot - iPhone X - 2020-03-04 at 09 45 46
You can't style things like text color in the picker itself, but that has nothing to do with the original View styling.

vonovak:
I'll release in the evening. Which reminds we need a changelog entry - would you be willing to write that @lochness42 ? Question now is, should the be a major or minor bump?

lochness42:
I would say minor as it fixes bug where height couldn't be readjusted, styling should be  preserved from before. So no real new behaviour apart from bug fix.

PR #124, opened 2020-03-03
Update Readme.md to fix example code

tz925:
# Summary
fixed Example code on readme.md so that the example code will not produce picker twice on Android, hence need to be confirmed/canceled twice.
I think this will help more confused people like myself, who copy paste the example code, and find it not working properly.

Area of impact, only in readme.md file.

## Checklist


tz925:
First time doing this, be easy on me please thank you 😂

luancurti:
@tz925 thanks for contribute!

PR #125, opened 2020-03-04
Update changelog/ios style property

lochness42:
# Summary

Changelog update and `package.json` version bump

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### 2.2.3
 
-- Fix iOS picker's styling bug where picker would be fixed to 216px height no matter what `style` was applied to the picker. Now it can correctly apply specific height settings and flex. NOTE: this only works with View style property type. [#120](https://github.com/react-native-community/react-native-datetimepicker/pull/120)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+### 2.2.3
+
+- Fix iOS picker's styling bug where picker would be fixed to 216px height no matter what `style` was applied to the picker. Now it can correctly apply specific height settings and flex. NOTE: this only works with View style property type. [#120](https://github.com/react-native-community/react-native-datetimepicker/pull/120)
```
vonovak:
```suggestion
- Fix iOS picker's styling bug where picker would be fixed to 216px height no matter what `style` was applied to the picker. This is because previously, the picker was always wrapped in a `View` which was now removed. Since `style` prop was never documented in this package, we do not consider this a breaking change. Now `style` can correctly apply specific height settings and flex. NOTE: this only works with View style property type. [#120](https://github.com/react-native-community/react-native-datetimepicker/pull/120)
```

ywriterct:
hi, wanted to check if this is working? I tried to set                  
 <DateTimePicker
    style={{ height: 100 }} ..... /> however the size of the date picker isn't changing
Screen Shot 2020-05-29 at 9 21 35 AM

Issue #20, opened 2019-08-31
Please add props to change theme color of the picker.


ajaykumar97:
# Feature Request

As the default picker theme is "White & green in android and White in iOS". I think there should be an option/prop to change the theme or color of the Date-Picker?

## Why it is needed

This feature is needed so as to match the picker color with the app theme and to make it more useful.

slorber:
+1 would be great!

Can it be done with a JS api only? or is it required to modify android theme files?

mwss1996:
+1, i realy love if it was implemented

jerryheir:
This color is the main issue why a lot of people don't use this package to be honest

robertherber:
+1

My main issue is not being able to change the gradient of the Android picker, not possible to make it look good on a non-white background.

guilleSequeiraWolox:
Hi, sorry my english is bad but i want to share how i can change theme color of the picker.
in `android/app/src/main/res/values/styles.xml` i putted:

```
  <style name="AppTheme" parent="Theme.AppCompat.Light.NoActionBar">
        <item name="android:datePickerDialogTheme">@style/DialogDatePicker.Theme</item>
    </style>

<style name="DialogDatePicker.Theme" parent="Theme.AppCompat.Light.Dialog">
        <item name="colorAccent">@color/ocean_green</item>
        <item name="android:textColor">@color/dove_gray</item>
        <item name="android:textColorPrimary">@color/dove_gray</item>
    </style>
``` 
- colorAccent is the color of header of calendar and selector
- textColor is the color of numbers and text buttons
- extColorPrimary is the color of text of month.


I hope it works for you, regards

jerryheir:
Thanks soooooo much

robertherber:
@guilleSequeiraWolox Thanks for the excellent example for how to style it statically. However - being able to set it dynamically is a must if your app supports any kinds of themes (which is a common case in these days of Dark Mode).

slorber:
Hi,

Unfortunately it will never be possible to change the color of this picker through a prop. The native widget does not allow so, and even android native developers complain about this.

The only way is through xml. If you want to have support for this,asking here won't help, you'd rather complain to google directly.

robertherber:
@slorber I see. Maybe it'd be possible to switch native android themes through this library? Then you'd at least be able to provide a dark and light option to switch between :)

mikehardy:
[...]

PR #127, opened 2020-03-06
Add textColor props for iOS datetimepicker

sbycrosz:
# Summary

Add textColor props for iOS datetimepicker (Solves the iOS part of #20)

## Test Plan

image

N/A

N/A

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    N/A     |

## Checklist

### DIFF
--- a/ios/RNDateTimePickerManager.m
+++ b/ios/RNDateTimePickerManager.m
@@ -40,12 +40,21 @@ RCT_EXPORT_VIEW_PROPERTY(onChange, RCTBubblingEventBlock)
 RCT_REMAP_VIEW_PROPERTY(mode, datePickerMode, UIDatePickerMode)
 RCT_REMAP_VIEW_PROPERTY(timeZoneOffsetInMinutes, timeZone, NSTimeZone)
 
-RCT_CUSTOM_VIEW_PROPERTY(textColor, NSNumber, RNDateTimePicker)
+RCT_CUSTOM_VIEW_PROPERTY(textColor, UIColor, RNDateTimePicker)
 {
-  if (!json) return;
-    
-  [view setValue:[RCTConvert UIColor:json] forKey:@"textColor"];
-  [view setValue:@NO forKey:@"highlightsToday"];
+  if (json) {
+    [view setValue:[RCTConvert UIColor:json] forKey:@"textColor"];
+    [view setValue:@(NO) forKey:@"highlightsToday"];
+  } else {
+    UIColor* defaultColor;
+    if (@available(iOS 13.0, *)) {
+        defaultColor = [UIColor labelColor];
+    } else {
+        defaultColor = [UIColor blackColor];
+    }
+    [view setValue:defaultColor forKey:@"textColor"];
+    [view setValue:@(YES) forKey:@"highlightsToday"];
+  }
 }
 
 @end
--- a/ios/RNDateTimePickerManager.m
+++ b/ios/RNDateTimePickerManager.m
@@ -40,8 +40,10 @@ RCT_EXPORT_VIEW_PROPERTY(onChange, RCTBubblingEventBlock)
 RCT_REMAP_VIEW_PROPERTY(mode, datePickerMode, UIDatePickerMode)
 RCT_REMAP_VIEW_PROPERTY(timeZoneOffsetInMinutes, timeZone, NSTimeZone)
 
-RCT_CUSTOM_VIEW_PROPERTY(textColor, UIColor, RNDateTimePicker)
+RCT_CUSTOM_VIEW_PROPERTY(textColor, NSNumber, RNDateTimePicker)
 {
+  if (!json) return;
+    
   [view setValue:[RCTConvert UIColor:json] forKey:@"textColor"];
   [view setValue:@NO forKey:@"highlightsToday"];
 }
--- a/ios/RNDateTimePickerManager.m
+++ b/ios/RNDateTimePickerManager.m
@@ -40,4 +40,10 @@ RCT_EXPORT_VIEW_PROPERTY(onChange, RCTBubblingEventBlock)
 RCT_REMAP_VIEW_PROPERTY(mode, datePickerMode, UIDatePickerMode)
 RCT_REMAP_VIEW_PROPERTY(timeZoneOffsetInMinutes, timeZone, NSTimeZone)
 
+RCT_CUSTOM_VIEW_PROPERTY(textColor, UIColor, RNDateTimePicker)
+{
+  [view setValue:[RCTConvert UIColor:json] forKey:@"textColor"];
+  [view setValue:@NO forKey:@"highlightsToday"];
+}
+
 @end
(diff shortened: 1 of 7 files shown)


src/types.js
```diff
@@ -99,6 +100,11 @@ export type IOSNativeProps = $ReadOnly<{|
    * instance, to show times in Pacific Standard Time, pass -7 * 60.
    */
   timeZoneOffsetInMinutes?: ?number,
+
+  /**
+   * The date picker text color.
+   */
+  textColor?: ColorValue,
```
vonovak:
```suggestion
  textColor?: ?ColorValue,
```

let's please have this optional

src/index.d.ts
```diff
@@ -87,6 +88,11 @@ export type IOSNativeProps = Readonly<BaseProps & {
    * instance, to show times in Pacific Standard Time, pass -7 * 60.
    */
   timeZoneOffsetInMinutes?: number,
+
+  /**
+   * The date picker text color.
+   */
+  textColor?: ColorValue,
```
vonovak:
```suggestion
  textColor?: ?ColorValue,
```
let's please have this optional

src/datetimepicker.ios.js
```diff
@@ -89,6 +90,7 @@ export default class Picker extends React.Component<IOSNativeProps> {
         minuteInterval={minuteInterval}
         timeZoneOffsetInMinutes={timeZoneOffsetInMinutes}
         onChange={this._onChange}
+        textColor={textColor}
```
vonovak: afaik we should first call `processColor` on this, for example see https://github.com/facebook/react-native/blob/ce0edca620d49597aa73c71a68b70efa170bc828/Libraries/Share/Share.js#L110

please also include the invariant check for color correctness, thank you

ios/RNDateTimePickerManager.m
```diff
@@ -40,4 +40,10 @@ - (UIView *)view
 RCT_REMAP_VIEW_PROPERTY(mode, datePickerMode, UIDatePickerMode)
 RCT_REMAP_VIEW_PROPERTY(timeZoneOffsetInMinutes, timeZone, NSTimeZone)
 
+RCT_CUSTOM_VIEW_PROPERTY(textColor, UIColor, RNDateTimePicker)
+{
+  [view setValue:[RCTConvert UIColor:json] forKey:@"textColor"];
```
vonovak: this function needs to be ready to be able to cope with `json` representing both the case when textColor is `ColorValue` and when it's undefined / null in which case it should use default color.

src/datetimepicker.ios.js
sbycrosz: I can't get it to work with processColor though. Any ideas?

sbycrosz: I'll investigate again tomorrow 😪

vonovak: @sbycrosz sorry, if that does not work then I was probably wrong. The conversion is probably done automatically since you specified it's a `UIColor` in `RCT_CUSTOM_VIEW_PROPERTY(textColor, UIColor, RNDateTimePicker)`

ios/RNDateTimePickerManager.m
```diff
@@ -40,4 +40,19 @@ - (UIView *)view
...
+  
+  if (@available(iOS 13.0, *)) {
+      defaultColor = [UIColor labelColor];
+  } else {
+      defaultColor = [UIColor blackColor];
+  }
+    
+  UIColor* textColor = json ? [RCTConvert UIColor:json] : defaultColor;
+  [view setValue:textColor forKey:@"textColor"];
+  [view setValue:@NO forKey:@"highlightsToday"];
```
vonovak: I'm wondering - what does this do? Should we set this to `@(YES)` if the color is specified?

src/datetimepicker.ios.js
sbycrosz: I managed to get this to work by using NSNumber instead of UIColor (as seen in https://github.com/facebook/react-native/issues/3014)

ios/RNDateTimePickerManager.m
sbycrosz: > what does this do?

This is to disable the automatic highlighting for today's date 

image

> Should we set this to @(YES) if the color is not specified?

Make sense! the setTextColor is now a noop when no color is specified

lochness42:
Does this still work correctly with dark/light mode? Also can we provide 2 values that will be applied depending which mode is currently active? That might be another PR though.

vonovak:
@lochness42 that will likely be handled at the `color` prop level and we don't need to worry about it, see https://github.com/facebook/react-native/pull/27908

ios/RNDateTimePickerManager.m
vonovak: sorry, I deleted this comment mistakenly, reposting

> the setTextColor is now a noop when no color is specified

this is in contrast to what was discussed before:

> this function needs to be ready to be able to cope with json representing both the case when textColor is ColorValue and when it's undefined / null in which case it should use default color.

^ this is so that when users render `DatePicker` with color=red and then re-render the component with color=undefined / null we need to change color back to default.

I didn't know what the default color was in `UIDatePicker` - it turns out it is `nil` or an actual color instance - at least on 13.3. Try running `UIColor* defaultColor = [defaultView valueForKey:@"textColor"];`. I think your previous implementation with `UIColor.labelColor` is the right way to go. For `highlightsToday` we should set the correct value too - NO when custom color is set, YES when no color is set. Meaning the implementation should look ~ like this:

```
RCT_CUSTOM_VIEW_PROPERTY(textColor, UIColor, RNDateTimePicker)
{
  UIColor* defaultColor;
  if (@available(iOS 13.0, *)) {
      defaultColor = [UIColor labelColor];
  } else {
      defaultColor = [UIColor blackColor];
  }
  UIColor* textColor = json ? [RCTConvert UIColor:json] : defaultColor;
  [view setValue:textColor forKey:@"textColor"];
  BOOL shouldHighlightToday = json;
  [view setValue:@(shouldHighlightToday) forKey:@"highlightsToday"];
}
```

does this make sense? Thanks.

vonovak:
@sbycrosz I'll polish the PR and merge now, so we get this done

pie6k:
image

any idea when it might be released to npm? :)

vonovak:
@pie6k it's there now

yepes:
guys, is it possible to use this without Xcode 11? im running 10.2 and i get:

No known class method for selector 'labelColor'

If I change Pods / development Pods / RNDateTimePicker RnDateTimePickerManager.m / @implementation RNDateTimePickerManager Line 51 to:
defaultColor = [UIColor blackColor];

it works on xcode 10 (the text will be always black, obviously). @sbycrosz  Anyway to make it to at least compile on olders xcode versions?

nenjamin2405:
Hi guys, thanks a lot for this PR, however, is there an option to change the color of the horizontal line of the. picker? Currently, I still see its default color is black.

PR #66, opened 2019-12-13
Android `minuteInterval` prop support

e1ferapontov:
# Summary

As we know, `minuteInterval` is only available for iOS now, and that's a feature many people desire to have on Android as well. These changes bring at least partial support of `minuteInterval` prop for Android.

Right now I've only covered 5 min increments of `minuteInverval` prop (mostly because I've only needed them for my work), but all the rest of iOS-only values should work nicely for `spinner` TimePicker mode with no or little changes.

## Test Plan
I've added additional `detox` test cases for this functionality.

## Checklist

- [X] I have tested this on a device and a simulator
- [X] I added the documentation in `README.md`
- [ ] I mentioned this change in `CHANGELOG.md`
- [X] I updated the typed files (TS and Flow)
- [X] I added a sample use of the API in the example project (`example/App.js`)

PR #166, opened 2020-05-15
Android `minuteInterval` prop support

BenderBRodrigez:
# Summary

Based on #66. Fixed Android build for RN 0.62

## Checklist


android/src/main/java/com/reactcommunity/rndatetimepicker/RNDismissableTimePickerDialog.java
```diff
@@ -31,28 +40,240 @@
  *   See: <a href="https://code.google.com/p/android/issues/detail?id=34833">Issue 34833</a>
  * </p>
  */
-public class RNDismissableTimePickerDialog extends TimePickerDialog {
+
+class CustomTimePickerDialog extends TimePickerDialog {
```
luancurti, line 44: Can you move this class to another java file?

android/src/main/java/com/reactcommunity/rndatetimepicker/RNTimePickerDialogFragment.java
```diff
@@ -55,9 +56,18 @@ static TimePickerDialog getDialog(
...
+      minuteInterval = args.getInt(RNConstants.ARG_INTERVAL);
+    }
+
     RNTimePickerDisplay display = RNTimePickerDisplay.DEFAULT;
     if (args != null && args.getString(RNConstants.ARG_DISPLAY, null) != null) {
-      display = RNTimePickerDisplay.valueOf(args.getString(RNConstants.ARG_DISPLAY).toUpperCase(Locale.US));
+      if (RNMinuteIntervals.isRadialPickerCompatible(minuteInterval)) {
+        display = RNTimePickerDisplay.valueOf(args.getString(RNConstants.ARG_DISPLAY).toUpperCase(Locale.US));
+      } else {
+        display = RNTimePickerDisplay.SPINNER;
```
luancurti, line 69: Why you are setting the display spinner? You can get from `args.getString(RNConstants.ARG_DISPLAY)`

android/src/main/java/com/reactcommunity/rndatetimepicker/RNDismissableTimePickerDialog.java
BenderBRodrigez: that's done

android/src/main/java/com/reactcommunity/rndatetimepicker/RNTimePickerDialogFragment.java
BenderBRodrigez: I'm not strong in java, I just fixed some issues from PR #66 to build this and get right necessary functionality in my project, according to the rule "works - don't touch it". In this case I don't understand what are you talking about. I'll be appreciated for more detailed explanation

luancurti: You are setting the `display` without respect the value of the `display` passed by the user is that right? I think you can check if the user passed the prop checking the `ARG_DISPLAY` if has value you se into `display` if not you set `SPINNER` what do you think?

PR #199, opened 2020-06-17
Support dark mode

### LINKED ISSUE #44: Date picker disappear
# Question

Platform: iOS 13.1.2, iPhone 6s, Xcode 11.1
Version: "@react-native-community/datetimepicker": "^2.1.0",
     "react-native": "0.61.2",
The date picker is not visible
image

but it works on  simulator with iOS 13.1
image

luancurti:
# Summary

If you change the background of the screen according dark mode the date time picker `disappear` because the text is black too, I've implemented the dark mode to change the text to white and background to black if dark mode is enabled

Closes #44

## Test Plan

Enable dark mode and the text should be white and the background should be black disable the dark mode and the text should be black and background is white

Test in Android:

Screenshot_1592948430

Test in iOS:

Simulator Screen Shot - iPhone 11 Pro - 2020-06-23 at 18 45 38

Enable dark mode on Android (API 29 or higher) and the date time picker `disappear` because the black text in black background

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    ✅     |

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,8 @@
     "react-native": ">=0.59"
   },
   "dependencies": {
-    "invariant": "^2.2.4"
+    "invariant": "^2.2.4",
+    "react-native-appstate-hook": "^1.0.4"
   },
   "optionalDependencies": {
     "react-native-windows": "^0.62.0-preview.1"
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "start:ios": "react-native run-ios",
     "start:windows": "react-native start --use-react-native-windows",
     "test": "jest ./test",
-    "posttest": "npm run lint",
+    "posttest": "yarn lint",
     "lint": "eslint {example,src,test}/**/*.js src/index.d.ts",
     "flow": "flow check",
     "detox:ios:build:debug": "detox build -c ios.sim.debug",
--- a/src/datetimepicker.ios.js
+++ b/src/datetimepicker.ios.js
@@ -9,12 +9,14 @@
  * @format
  * @flow strict-local
  */
+import React, {useEffect} from 'react';
 import {StyleSheet} from 'react-native';
+import {Colors} from 'react-native/Libraries/NewAppScreen';
+import invariant from 'invariant';
+
 import RNDateTimePicker from './picker';
 import {toMilliseconds} from './utils';
 import {MODE_DATE} from './constants';
-import invariant from 'invariant';
-import React from 'react';
 
 import type {
   Event,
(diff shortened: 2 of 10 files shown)


danieloprado:
I've tested on my iPhone, the option isDarkModeEnabled need some changes, but with false now I can use it with dark mode on.

isDarkModeEnabled=true
WhatsApp Image 2020-06-23 at 13 21 32
isDarkModeEnabled=false
WhatsApp Image 2020-06-23 at 13 22 27

src/datetimepicker.ios.js
```diff
@@ -29,71 +31,68 @@ const styles = StyleSheet.create({
...
+  testID,
+  minuteInterval,
+  timeZoneOffsetInMinutes,
+  textColor,
+  onChange,
+  isDarkModeEnabled,
+}: IOSNativeProps) {
+  const dates: DatePickerOptions = {value, maximumDate, minimumDate};
+  const picker: NativeRef = React.createRef();
+  const textColorByDarkModeEnabled = isDarkModeEnabled
```
vonovak, line 50: this does not actually do anything - you just recreated the default behavior.

when dark mode is on, text is white,

when dark mode is off, text is black 

so this will not have any effect - what DOES have effect is `const backgroundColorByMode = isDarkModeEnabled ? Colors.black : Colors.white;` that you have added to `example/App.js`

src/datetimepicker.ios.js
```diff
@@ -9,12 +9,14 @@
  * @format
  * @flow strict-local
  */
+import React, {useEffect} from 'react';
 import {StyleSheet} from 'react-native';
+import {Colors} from 'react-native/Libraries/NewAppScreen';
```
vonovak, line 14: I would rather not rely on these as I do not consider these to be part of the stable api

package.json
```diff
@@ -75,7 +75,8 @@
     "react-native": ">=0.59"
   },
   "dependencies": {
-    "invariant": "^2.2.4"
+    "invariant": "^2.2.4",
+    "react-native-appstate-hook": "^1.0.4"
```
vonovak, line 79: this would be a development dep only, because it is not used by the library itself but just the example app

vonovak:
@luancurti here's an alternative way without a new prop https://github.com/react-native-community/datetimepicker/pull/204

luancurti:
> @luancurti here's an alternative way without a new prop #204

@vonovak I liked this implementation, sounds more clear

PR #784, opened 2023-07-11
fix(android): gradle 8 compatibility

CoreyCole:
fixes react-native-datetimepicker#783 React native 72 upgrades gradle from 7 to 8. This patch resolved the following issue:
```
Reason: Task ':@react-native-community_datetimepicker:compileDebugJavaWithJavac' uses this output of task ':react-native-community_datetimepicker:compileDebugAidl' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.
```

# Summary

Upgrading my project from RN 71 -> 72 (gradle 8) caused a build failure from this datetimepicker dependency. I was able to fix the build failure by adding a patch to `node_modules/@react-native-community/datetimepicker/android/build.gradle`.

Build for android on react native 72 & gradle 8

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅      |

## Checklist

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -62,19 +62,6 @@ dependencies {
     implementation 'com.facebook.react:react-native:+'
 }
 
-project.afterEvaluate {
-    def packageDebugResources = project.tasks.findByName('react-native-community_datetimepicker:packageDebugResources')
-    def compileDebugRenderscript = project.tasks.findByName('compileDebugRenderscript')
-    def generateDebugResValues = project.tasks.findByName('generateDebugResValues')
-
-    if (packageDebugResources != null && compileDebugRenderscript != null && generateDebugResValues != null) {
-        packageDebugResources.dependsOn compileDebugRenderscript, generateDebugResValues
-    }
-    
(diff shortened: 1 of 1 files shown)


CoreyCole:
This is very fragile though for a number of reasons:

- Anyone wanting to use your code now also needs to use afterEvaluate {} which is not obvious at all
- Anyone wanting to use the code that uses your code also needs to use afterEvaluate{} and so on, all the way down
- It becomes even more complicated as more plugins/scripts are involved
- More generally, it makes dependencies between plugins/scripts implicit and error prone

https://blog.mbonnin.net/my-life-after-afterevaluate

There might be a better way to make `:compileDebugAidl` finish before `: compileDebugJavaWithJavac` without using afterEvaluate?

```groovy
project.tasks.named('compileDebugJavaWithJavac').configure { compileDebugJavaWithJavac ->
    compileDebugJavaWithJavac.dependsOn project.tasks.named('compileDebugAidl')
}
```

PR #830, opened 2023-10-19
Fix incorrect UIDatePicker size computation when UIDatePickerModeDateAndTime + UIDatePickerStyleInline

mattsellars123:
# Summary

Fix issue where UIDatePicker sizeThatFits: doesn't account for the bottom 44 height of the time picker row when configured with datePickerMode == UIDatePickerModeDateAndTime and datePickerStyle == UIDatePickerStyleInline. There are a few examples online of others running into this where wrapping in a parent view and configuring autolayout to the parent anchors fixes the issue but that did not seem like a proper fix for how UIDatePicker is used here. I've isolated this change ONLY to this use case as other calls to get size appear correct.

* What areas of the library does it impact?

This impacts the size calculation which is provided to react.

## Test Plan
Pre-fix: IMG_2333

Post-fix: CleanShot 2023-10-18 at 23 13 33@2x

Debugging the size's height would always come out to about 348 when it should be closer to 392 in order to account for the bottom time selection row

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅      |
| Android |    ❌     |

## Checklist

### DIFF
--- a/ios/RNDateTimePickerShadowView.m
+++ b/ios/RNDateTimePickerShadowView.m
@@ -70,6 +70,15 @@ static YGSize RNDateTimePickerShadowViewMeasure(YGNodeRef node, float width, YGM
 
     size = [shadowPickerView.picker sizeThatFits:UILayoutFittingCompressedSize];
     size.width += 10;
+    if (@available(iOS 14.0, *)) {
+      if(shadowPickerView.picker.datePickerMode == UIDatePickerModeDateAndTime
+         && shadowPickerView.picker.datePickerStyle == UIDatePickerStyleInline) {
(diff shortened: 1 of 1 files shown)


vonovak:
Hello and thanks for the PR,
please see the example project screenshots - this seems to break the UI in our example project case. 

<details>
  <summary>before and after</summary>
  
Simulator Screenshot - iPhone 14 Pro - 2023-10-23 at 13 14 43

Simulator Screenshot - iPhone 14 Pro - 2023-10-23 at 13 14 56

</details>

Feel free to tinker with this more. The problem might be in your project, in our example project, or in the library.
But at this point I don't think this is good to be merged.
Thank you 🙂

PR #913, opened 2024-07-03
Improve color resource handling in Common.java

andvalsol:
The changes handle the fetching of color resources more accurately in our Common.java file. The old method was simplified to one line, but this didn't consider raw color values. Now, the code correctly handles both cases: when the type value is resolved to a color resource ID, and when it's resolved to a raw color value.

# Summary

Handling Default Values
If the textColorPrimary attribute is not explicitly defined in your theme, the typedValue.resourceId might still be 0, and typedValue.data might contain a raw color value instead of a resource ID. This could lead to unexpected behavior if you're expecting a resource ID.
Explicit Check: We now explicitly check if typedValue.resourceId is non-zero before assuming it's a resource ID.
Raw Color Handling: If typedValue.resourceId is 0, we directly use typedValue.data as the color value since it likely represents a raw color.
This modification ensures that your code gracefully handles cases where the resolved attribute value is either a color resource ID or a raw color value.

N/A

If there's no default default text color in the App's theme the app that implements this library will crash.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅ ❌     |
| Android |    ✅ ❌     |

## Checklist

### DIFF
--- a/android/src/main/java/com/reactcommunity/rndatetimepicker/Common.java
+++ b/android/src/main/java/com/reactcommunity/rndatetimepicker/Common.java
@@ -63,8 +63,14 @@ public class Common {
 		TypedValue typedValue = new TypedValue();
 		Resources.Theme theme = activity.getTheme();
 		theme.resolveAttribute(android.R.attr.textColorPrimary, typedValue, true);
-		@ColorRes int colorRes = (typedValue.resourceId != 0) ? typedValue.resourceId : typedValue.data;
-		@ColorInt int colorId = ContextCompat.getColor(activity, colorRes);
+    @ColorInt int colorId;
+    if (typedValue.resourceId != 0) {
+      // Resolved to a color resource ID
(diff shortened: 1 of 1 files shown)


vonovak (changes requested):
can you please 
Hello, and thank you for the PR!

Can you please fix the indentation?

Thank you 👍

andvalsol:
@vonovak Hello, I tried to see if there were any identation errors but according to my Android Studio the formatting is correct, could you please tell which identation format are you using?

vonovak:
@andvalsol I'm afraid that different parts of the code base are formatted differently so there isn't a unified formatting, but if you look at the diff tab here in github you'll see it. Just follow the existing indentation and we're fine :)

TY!

Screenshot 2024-07-09 at 9 58 01

vonovak (approved):

vonovak:
:tada: This issue has been resolved in version 8.2.0 :tada:

Issue #907, opened 2024-06-10
DateTimePicker Error


eypcnmrtOPT:
# Question

When using the time and date picker in react native, the date is selected but the time cannot be selected. When date is selected “Warning: Internal React error: Attempted to capture a commit phase error inside a detached tree. This indicates a bug in React. Likely causes include deleting the same fiber more than once, committing an already-finished tree, or an inconsistent return pointer.

Error message:

%s%s, TypeError: Cannot read property 'dismiss' of undefined, 
    in RNDateTimePickerAndroid” this error appears. Neither the time picker opens nor the date picker closes.

Icheka:
I have this problem too.

brandon-austin-lark:
same issue here. any updates?

azizmobarak:
confirmed

MuhammetFurkanOzenn:
any updates?

brandon-austin-lark:
Also getting this error logged

CarlosX4vier:
I have this problem too. Any solution?

butterfly-valley:
I have this problem too

ianemv:
Confirmed! Also having same issue

fimbres:
Same here

srs1234peng:
Same here

SimantaRajSarma:
same here :)

vishwa-rn:
```
type IOSMode = 'date' | 'time' | 'datetime' | 'countdown';
type AndroidMode = 'date' | 'time';

```
The above code in the following file
`client/node_modules/@react-native-community/datetimepicker/src/index.d.ts`

The following will work only in ios. 
```
<DateTimePicker
            value={paymentDate || new Date()}
            mode="datetime"
            display={Platform.OS === "ios" ? "spinner" : "default"}
            onChange={handlePaymentDateChange}
            style={{ width: "100%" }}
          />
```

For android, you need to create two pickers - one for date and other for time and handle the date time separately.
```
<DateTimePicker
            value={paymentDate || new Date()}
            mode="date"
            display="default"
            onChange={handlePaymentDateChange}
            style={{ width: "100%" }}
          />

<DateTimePicker
            value={paymentDate || new Date()}
            mode="time"
            display="default"
            onChange={handlePaymentTimeChange}
            style={{ width: "100%" }}
          />
```

perryfaro:
[...]

PR #929, opened 2024-09-10
Fixed issue #907 TypeError: Cannot read property 'dismiss' of undefined

herberthk:
# Summary
The error was described in [this issue](https://github.com/react-native-datetimepicker/datetimepicker/issues/907)

* What issues does the pull request solve? it solves [this issue](https://github.com/react-native-datetimepicker/datetimepicker/issues/907)

### What's required for testing? run the code below on android
```ts
                 <DateTimePicker
                        testID="dateTimePicker"
                        timeZoneOffsetInMinutes={0}
                        value={date}
                        mode={mode}
                        is24Hour={true}
                        display={Platform.OS === 'ios' ? 'inline' : 'default'}
                        onChange={onChange}
                        minimumDate={new Date()}
                    />
```

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |     ❌     |
| Android |    ✅      |

## Checklist

### DIFF
--- a/src/DateTimePickerAndroid.android.js
+++ b/src/DateTimePickerAndroid.android.js
@@ -118,7 +118,7 @@ function open(props: AndroidNativeProps) {
 
 function dismiss(mode: AndroidNativeProps['mode']): Promise<boolean> {
   // $FlowFixMe - `AbstractComponent` [1] is not an instance type.
-  return pickers[mode].dismiss();
+  return pickers[mode]?.dismiss();
 }
 
 export const DateTimePickerAndroid = {open, dismiss};


src/DateTimePickerAndroid.android.js
```diff
@@ -118,7 +118,7 @@ function open(props: AndroidNativeProps) {
 
 function dismiss(mode: AndroidNativeProps['mode']): Promise<boolean> {
   // $FlowFixMe - `AbstractComponent` [1] is not an instance type.
-  return pickers[mode].dismiss();
+  return pickers[mode]?.dismiss();
```
vonovak, line 121: hello and thanks for opening the PR.
I'm sure this fixes the error of "Cannot read property 'dismiss' of undefined" but I'd like to know first why `pickers[mode]` is undefined. It shouldn't be, as far as I can tell.
In the issue that you linked there's a lot of comments, but no one was able to provide a reproduction of the issue.

Right now, I see this more as a workaround rather than a solution. I'd like to know better what the underlying problem is in order to be able to assess if this is the right fix.

Can you provide more insight into what's happening?
Thank you

herberthk:
@vonovak Thanks for your response and contribution to this project. I'm not sure why pickers[mode] is undefined but this error came after migrating to react-native 0.75.2, I upgraded to the latest version of this package thinking it will resolve the issue which was not the case. The error comes after selecting or picking the date.

The error on the app looks like this
Screenshot_20240913_125231_Connect Up

The error in the terminal looks like this
Screenshot from 2024-09-13 12-54-25

This is my system information
Screenshot from 2024-09-13 13-02-20

maiznadeem:
When can this be resolved? Or can I downgrade my version?

herberthk:
@maiznadeem The maintainer is not responding yet this PR fixes the issue

vonovak:
Hello,
I have responded in the review comment: https://github.com/react-native-datetimepicker/datetimepicker/pull/929#discussion_r1756341631

Right now, I don't have the confidence that this fix is the correct fix. More investigation should be done to determine the cause of the issue, and then assess if this is the correct fix or not.

sanchaz:
@maiznadeem see https://github.com/react-native-datetimepicker/datetimepicker/issues/907#issuecomment-2301571623
The issue seems to be that when choosing 'datetime' mode on android, the picker for date still appears.
If I run only date and time (I know this is documented), it works fine.
So no fixed needed, but to avoid these comments/issues... maybe throwing without even loading the picker?

maiznadeem:
> @maiznadeem see https://github.com/react-native-datetimepicker/datetimepicker/issues/907#issuecomment-2301571623
> The issue seems to be that when choosing 'datetime' mode on android, the picker for date still appears.
> If I run only date and time (I know this is documented), it works fine.
> So no fixed needed, but to avoid these comments/issues... maybe throwing without even loading the picker?

Just realised! thanks I might need to make my own datetime picker now.

sanchaz:
> > @maiznadeem see [#907 (comment)](https://github.com/react-native-datetimepicker/datetimepicker/issues/907#issuecomment-2301571623)
> > The issue seems to be that when choosing 'datetime' mode on android, the picker for date still appears.
> > If I run only date and time (I know this is documented), it works fine.
> > So no fixed needed, but to avoid these comments/issues... maybe throwing without even loading the picker?
> 
> Just realised! thanks I might need to make my own datetime picker now.

@maiznadeem the way it is usually done on android, if you look at google calendar for example, is that you show the date picker first and then the time picker as soon as the date picker returns. I would recommend following that pattern instead of rolling your own datetime picker.

maiznadeem:
> > 
> > Just realised! thanks I might need to make my own datetime picker now.
> 
> @maiznadeem the way it is usually done on android, if you look at google calendar for example, is that you show the date picker first and then the time picker as soon as the date picker returns. I would recommend following that pattern instead of rolling your own datetime picker.

Haha, no I also meant searching for one like Tamagui or somewhere.

I could go with paper one, but it's such a pain, when handling validations, for example I want to select a date and time that is 3 hours from now. But the component returns me datetime values in UTC. I first need to extract time and date from each datetimes. And for example, in my current timezone, 3 hours from now might be the next day, but in UTC it's not, so validation also fails there.

fraction01:
I also have this issue while using mode="datetime" on android, I expect the date picker to be displayed first, then the time picker

Griffendoors:
Hi all,  I ran into this issue and found a solution - commenting in case it can help others.

The 'TypeError: Cannot read property 'dismiss' of undefined' occurs when trying to use mode = 'datetime' with Android, which is not supported.

Screenshot 2026-06-03 at 4 54 52 pm

To provide functionality for a date time picker in Android, you can conditionally render one and then the other, which is working well for my team.

```ts
export interface Props {
    value?: Date;
    onValueChange: (date: Date) => void;
}

const [step, setStep] = useState<'date' | 'time'>('date');
const [date, setDate] = useState<Date | null>(null);

if (step === 'date') {
    return (
        <DateTimePicker
            value={props.value}
            mode={'date'}
            display={'default'}
            onValueChange={(_e, date) => {
                setDate(date || null);
                setStep('time');
            }}
        />
    );
} else if (step === 'time') {
    return (
        <DateTimePicker
            value={props.value}
            mode={'time'}
            display={'default'}
            onValueChange={(_e, time) => {
                if (date && time) {
                    const combinedDateTime = new Date(
                        date.getFullYear(),
                        date.getMonth(),
                        date.getDate(),
                        time.getHours(),
                        time.getMinutes(),
                        time.getSeconds(),
                    );
                    props.onValueChange(combinedDateTime);
                }
                setStep('date');
                setDate(null);
            }}
        />
    );
}


```

PR #973, opened 2025-05-12
Bump react version

Phil9l:
Update react to make it compatible with a new expo version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
   "peerDependencies": {
     "expo": ">=50.0.0",
     "react": "^19.0.0",
-    "react-native": "^0.79.2"
+    "react-native": "^0.79.2",
     "react-native-windows": "*"
   },
   "peerDependenciesMeta": {


rdundon:
Came across this too as I noticed React 19 is currently not supported and was looking if there were any existing PRs.

Is bumping the `package.json` file the only thing needed? Also, looks like some CI builds are failing. Were you able to get this running locally?

Samide47:
why this is PR not merged yet because the package does not work properly with RN 0.79 on Android

rdundon:
> why this is PR not merged yet because the package does not work properly with RN 0.79 on Android 

Yeah, but does this PR actually work/fix the issue?

Looks like one of the CI builds failed because the 'package-lock.json' isn't in sync. That would at least need to be addressed. But I'm not sure if there's any thing else that may have been missed.

I suppose if time allows I could try running locally and seeing if the PR does indeed work. 

Not the maintainer, but also interested in seeing React 19 compatibility resolved

Ali-Aref:
Any updates ?

hraschan:
When will this be fixed? 

Cant upgrade to expo 53

PR #1009, opened 2025-09-10
fix(ios): avoid min>max date picker crash

sterlingwes:
# Summary

This change fixes a crash on iOS when the underlying UIDatePicker receives minimumDate > maximumDate by:

- adding validation in JS to assert that when both are set, min must be less than max
- updating the native fabric prop update logic to:
  - explicitly handles unsetting prior values on the underlying picker during any change before setting new ones
  - specifically handles the unset case which was handled implicitly before

This is the error that shows after this change in dev mode when this happens (before it just crashes):

Simulator Screenshot - iPhone 16 Pro - 2025-09-09 at 21 58 50

<details>
<summary>Original Crash</summary>

```
*** Terminating app due to uncaught exception 'NSGenericException', reason: 'Start date cannot be later in time than end date!'
*** First throw call stack:
(
	0   CoreFoundation                      0x00000001804c9690 __exceptionPreprocess + 172
	1   libobjc.A.dylib                     0x00000001800937cc objc_exception_throw + 72
	2   Foundation                          0x0000000180f55184 -[_NSConcreteDateInterval initWithStartDate:endDate:] + 396
	3   UIKitCore                           0x00000001850dfee0 -[_UIDatePickerCalendarView _reloadCalendarView] + 352
	4   UIKitCore                           0x00000001850dffa4 -[_UIDatePickerCalendarView _reload] + 48
	5   UIKitCore                           0x0000000185bfd72c -[UIDatePicker _installPickerView:updatingSize:] + 152
	6   UIKitCore                           0x0000000185bfd650 -[UIDatePicker _updatePickerViewIfNecessary] + 120
	7   ReproducerApp.debug.dylib           0x0000000107242670 -[RNDateTimePickerComponentView updatePropsForPicker:props:oldProps:
```

</details>

## Test Plan

Reproducer used in screenshots below: https://github.com/sterlingwes/date-picker-min-max-repro

The reproducer tests the following scenarios:

Simulator Screenshot - iPhone 16 Pro - 2025-09-09 at 19 21 13

<details>
<summary>Behaviour before fixes</summary>

| first | second                                                                                                                                                                                       | third                                                                                                                                                                                        | fourth                                                                                                                                                                                       | fifth                                                                                                                                                                                        |
[...]

PR #1026, opened 2025-11-23
fix(ios): internal picker state persisting with recycling

sterlingwes:
# Summary

Follow-up to the min/max constraint fix in https://github.com/react-native-datetimepicker/datetimepicker/pull/1009. There was a pre-existing issue when the component gets unmounted: the native component gets recycled for future reuse but keeps its picker instance properties along with their state.

This likely would have also fixed the earlier issue but I think those changes are worth keeping as they make it clearer how the UIDatePicker constraints operate.

## Test Plan

**Video showing the crash case:**

https://github.com/user-attachments/assets/da84a496-3278-47e4-9b85-f109862a2bb5

Two breakpoints here:
* one in the recycle lifecycle method to show that the picker retains constraints
* one right before we apply props for the next picker to show that the constraints are stale before the crash

**Video showing the fix:**

https://github.com/user-attachments/assets/af535623-ce43-4287-ab88-b2657f08ec06

Breakpoints show that the min/max date constraints are cleared

I copied over [this component from the reproduction app](https://github.com/Junveloper/react-native-datetimepicker-bug/blob/main/app/date-modal.tsx) ([provided by @Junveloper in the prior PR](https://github.com/react-native-datetimepicker/datetimepicker/pull/1009#issuecomment-3568011729)). You can reference this branch for the changes required: https://github.com/sterlingwes/datetimepicker/pull/1

* open the modal
  * pick two dates in the next month
  * submit / close modal
* open modal again
  * tap to pick first date and it should crash

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

- [x] I have tested this on a simulator

### DIFF
--- a/ios/fabric/RNDateTimePickerComponentView.mm
+++ b/ios/fabric/RNDateTimePickerComponentView.mm
@@ -115,10 +115,9 @@ NSDate* adjustMinimumDate (NSDate* minimumDate, int minuteInterval) {
     return concreteComponentDescriptorProvider<RNDateTimePickerComponentDescriptor>();
 }
 
-- (void)prepareForRecycle
++ (BOOL)shouldBeRecycled
 {
-    [super prepareForRecycle];
-    _state.reset();
+  return NO;
 }
 
 -(void)updateTextColorForPicker:(UIDatePicker *)picker color:(UIColor *)color


ios/RNDateTimePicker.m
```diff
@@ -72,4 +72,11 @@ - (void)setDate:(NSDate *)date {
     }
 }
 
+- (void)reset
+{
+  self.minimumDate = nil;
+  self.maximumDate = nil;
+  self.date = [NSDate date];
```
sterlingwes: can't set this to nil - this appears to be the default when you instantiate a UIDatePicker

Junveloper:
Thank you @vonovak  and @sterlingwes . When could I expect a new release? :)

PR #1031, opened 2025-12-24
fix(android): material date picker off by one day

mjmasn:
# Summary
The Material DatePicker returns a midnight UTC timestamp when picking a date, so when the timezone was behind UTC (e.g. PST), the final date would be set to the previous day to the one picked.

## Test Plan

Tested in Android Emulator set to PST timezone

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    n/a     |
| Android |    ✅     |

## Checklist

- [x] I have tested this on a ~device and a~ simulator

### DIFF
--- a/android/src/main/java/com/reactcommunity/rndatetimepicker/RNMaterialDatePicker.kt
+++ b/android/src/main/java/com/reactcommunity/rndatetimepicker/RNMaterialDatePicker.kt
@@ -179,8 +179,14 @@ class RNMaterialDatePicker(
         )
       )
 
-      newCalendar.timeInMillis = selection
+      // Material DatePicker returns timestamp in UTC at midnight for the selected date.
+      // Extract year, month, day from UTC, then set them in the target timezone
+      val utcCalendar = Calendar.getInstance(java.util.TimeZone.getTimeZone("UTC"))
+      utcCalendar.timeInMillis = selection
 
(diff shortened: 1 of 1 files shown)


mjmasn:
I just mean a timestamp equivalent to the user-selected date at 00:00:00.000Z, which refers to the previous day in timezones behind UTC (e.g. anything in USA)

Repro steps:

1. Add a date/time picker to your app with the `design="material"` prop (there's other setup to make material pickers work IIRC, I guess follow the documentation...)
2. Run the app in android on a device or emulator set to PST timezone (or any behind-UTC timezone)
3. Select a date and time
4. Returned day of month is 1 day behind your selection, but with the correct local time. 

The only workaround is to select the wrong date initially, which users understandably aren't too keen on.

The fix is reading the timestamp from the material date picker as UTC then extracting the UTC year/month/day of month and setting those values back onto the local time so that the day of month the user picks is the one selected, regardless of the user's timezone offset from UTC.

It seems the author assumed the timestamp returned was midnight in the local time on the selected date, which was probably a fair assumption if the old style picker worked that way, but unfortunately this is not the case with the material picker.

We've patched this in our app already and it's now working correctly, but hopefully this helps anyone else who has encountered the same issue.

ferrannp:
Thank you a lot!

